PR #56, opened 2020-05-31
Looped conditional imports aren't resolved properly

### LINKED ISSUE #57: Nested-scope (e.g. conditional) imports in files extended from loop are not resolved
As requested in #56, this is the issue tracking the bug, that imports aren't resolved when they happen in certain nested scopes inside the imported file.

The two test cases I've written in #56 are here:

```swift
func testLoopedConditionalImport() throws {
    var test = TestFiles()
    test.files["/base.leaf"] = """
    #for(x in list):
    #extend("entry"):#export("something", "Whatever")#endextend
    #endfor
    """
    test.files["/entry.leaf"] = """
    #(x): #if(isFirst):#import("something")#else:Not First#endif
    """

    let expected = """

    A: Whatever

    B: Not First

    C: Not First

    """

    let renderer = LeafRenderer(
        configuration: .init(rootDirectory: "/"),
        cache: DefaultLeafCache(),
        files: test,
        eventLoop: EmbeddedEventLoop()
    )

    let page = try renderer.render(path: "base", context: ["list": ["A", "B", "C"]]).wait()
    XCTAssertEqual(page.string, expected)
}

func testMultipleLoopedConditionalImports() throws {
    var test = TestFiles()
    test.files["/base.leaf"] = """
    #for(x in list1):
    #extend("entry"):#export("something", "Whatever")#endextend
    #endfor
    #for(x in list2):
    #extend("entry"):#export("something", "Something Else")#endextend
    #endfor
    """
    test.files["/entry.leaf"] = """
    #(x): #if(isFirst):#import("something")#else:Not First#endif
    """

    let expected = """

    A: Whatever

    B: Not First

    C: Not First


    A: Something Else

    B: Not First

    C: Not First

    """

    let renderer = LeafRenderer(
        configuration: .init(rootDirectory: "/"),
        cache: DefaultLeafCache(),
        files: test,
        eventLoop: EmbeddedEventLoop()
    )

    let page = try renderer.render(path: "base", context: [
        "list1": ["A", "B", "C"],
        "list2": ["A", "B", "C"],
    ]).wait()
    XCTAssertEqual(page.string, expected)
}
```

Furthermore, `imports` might appear in all kinds of different places (e.g. inside parameters of another tag).
One thing I noticed in my approach to it, is that care has to be taken with `Conditional` since it's a class. My first approach was not resolving imports in repetitive `extend`s correctly anymore, because the `Conditional` in the included file was only created once and thus only resolved in the first pass. The second `embed` would show the `export`s of the first `extend`. This is covered by the second test-case, though.

ffried:
When using `#import` inside an `#if` that is nested inside a file that is embedded using `#extend`, the rendering fails since apparently the `#import` isn't resolved.

I've tried to find the cause and fix it myself, but wasn't able to understand the process in reasonable time. So I've at least added a (currently failing) test for it.

Resolves #57

### DIFF
--- a/Sources/LeafKit/LeafAST.swift
+++ b/Sources/LeafKit/LeafAST.swift
@@ -48,7 +48,7 @@ public struct LeafAST: Hashable {
                 case .extend(let e):
                     let key = e.key
                     if let insert = externals[key] {
-                        let inlined = e.extend(base: insert.ast)
+                        let inlined = insert.ast.resolvingImports(exports: e.exports)
                         ast.replaceSubrange(pos...pos, with: inlined)
                     } else {
                         unresolvedRefs.insert(key)
--- a/Sources/LeafKit/LeafLexer/LeafToken.swift
+++ b/Sources/LeafKit/LeafLexer/LeafToken.swift
@@ -296,6 +296,19 @@ internal extension Array where Element == ParameterDeclaration {
         }
         return nil
     }
+
+    func resolvingImports(exports: [String: Syntax.Export]) -> [ParameterDeclaration] {
+        map {
+            switch $0 {
+            case .parameter(_): return $0
+            case .expression(let params):
+                return .expression(params.resolvingImports(exports: exports))
+            case .tag(var customTag):
+                customTag.resolveImports(exports: exports)
+                return .tag(customTag)
+            }
+        }
+    }
     
     func describe(_ joinBy: String = " ") -> String {
         return self.map {$0.short }.joined(separator: joinBy)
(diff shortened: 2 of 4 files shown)


ffried:
I've given it another shot and actually found a solution that turns my test green. I'm not sure this is the optimal solution though, since it needs to iterate the AST and I'm not sure if all of these iterations are necessary.
If there's a better solution, I'd have no problem with my second commit being replaced with something better. 🙂

tdotclare:
Thanks for posting this! Can you open a corresponding issue with just your test statement for tracking resolution?

I'm looking through it now and will have a separate PR fixing this myself as it needs to be somewhat more generalized; after reviewing the issue it occurs to me that it needs to be more substantial since there's an implied possibility that one might #export/#import tag parameters as well and it's necessary to evaluate that possibility as well.

ffried:
@tdotclare Done -> #57 

I was pretty sure that my approach to it wouldn't be the one that makes it. 😅
However, I needed a quick solution to continue migrating to Leaf 4. Once I found one, I thought I'd push it nevertheless. I've also written down some notes in #57 of what I found to be some challenges when resolving imports.

tdotclare:
Closing as superseded by PR #58

PR #58, opened 2020-06-02
Improved #export/#import Behavior

tdotclare:
This update provides significantly improved resolution of #export/#import.

### API Changes
The unbodied usage of `#export("exportkey", value)` now supports any acceptable single parameter as the exported value. Example uses:
```swift
// Exports context variable
#export("isAdmin", admin)
// Exports an expression
#export("specialKey", username == "tdotclare")
// Exports a custom tag call
#export("baseURL", baseURL()) 
```
`#import` can now be used inside parameters, allowing behavior like:
```swift
#if(import("showSidebar")): <div id="sidebar"><div>  #endif
```
### Observable Behavior Changes/Fixes
* Fixes Issue #57 - `#import` syntaxes are now correctly replaced when nested inside other objects like loops

### Internal Changes
* `Syntax.Conditional` is now a chained-block struct instead of linked classes
* Adds `BodiedSyntax` protocol to applicable `Syntax`es, replacing monolithic extensions
  * Publish & maintain external or import dependency sets for adherants
  * Provide `inlineRefs` methods to adherents to self-rebuild
* Smarter `LeafAST` resolution
* Provides applicable `import` capabilities to ParameterDescription and Parameter

### DIFF
--- a/Sources/LeafKit/LeafAST.swift
+++ b/Sources/LeafKit/LeafAST.swift
@@ -44,21 +44,30 @@ public struct LeafAST: Hashable {
 
         // inline provided externals
         while pos < ast.endIndex {
-            switch ast[pos] {
-                case .extend(let e):
-                    let key = e.key
-                    if let insert = externals[key] {
-                        let inlined = e.extend(base: insert.ast)
-                        ast.replaceSubrange(pos...pos, with: inlined)
-                    } else {
-                        unresolvedRefs.insert(key)
-                        pos = ast.index(after: pos)
-                    }
-                default:
-                    var new: Syntax? = nil
-                    unresolvedRefs.formUnion(ast[pos].inlineRefs(externals, &new))
-                    if let new = new { ast[pos] = new }
-                    pos = ast.index(after: pos)
+            // get desired externals for this Syntax - if none, continue
+            let wantedExts = ast[pos].externals()
+            if wantedExts.isEmpty {
+                pos = ast.index(after: pos)
+                continue
+            }
+            // see if we can provide any of them - if not, continue
+            let providedExts = externals.filter { wantedExts.contains($0.key) }
+            if providedExts.isEmpty {
+                unresolvedRefs.formUnion(wantedExts)
+                pos = ast.index(after: pos)
+                continue
+            }
+            
+            // replace the original Syntax with the results of inlining, potentially 1...n
+            let replacementSyntax = ast[pos].inlineRefs(providedExts, [:])
+            ast.replaceSubrange(pos...pos, with: replacementSyntax)
+            // any returned new inlined syntaxes can't be further resolved at this point
+            // but we need to add their unresolvable references to the global set
+            var offset = replacementSyntax.startIndex
+            while offset < replacementSyntax.endIndex {
+                unresolvedRefs.formUnion(ast[pos].externals())
+                offset = replacementSyntax.index(after: offset)
+                pos = ast.index(after: pos)
             }
         }
 
--- a/Sources/LeafKit/LeafLexer/LeafToken.swift
+++ b/Sources/LeafKit/LeafLexer/LeafToken.swift
@@ -203,7 +203,7 @@ public indirect enum ParameterDeclaration: CustomStringConvertible {
         }
     }
     
-    func imports() -> Set<String> {
+    internal func imports() -> Set<String> {
         switch self {
             case .parameter(_): return .init()
             case .expression(let e): return e.imports()
@@ -212,12 +212,12 @@ public indirect enum ParameterDeclaration: CustomStringConvertible {
                 guard let parameter = t.params.first,
                       case .parameter(let p) = parameter,
                       case .stringLiteral(let key) = p,
-                      key.count > 0 else { return .init() }
+                      !key.isEmpty else { return .init() }
                 return .init(arrayLiteral: key)
         }
     }
     
-    func inlineImports(_ imports: [String : Syntax.Export]) -> ParameterDeclaration {
+    internal func inlineImports(_ imports: [String : Syntax.Export]) -> ParameterDeclaration {
         switch self {
             case .parameter(_): return self
             case .tag(let t):
@@ -227,7 +227,6 @@ public indirect enum ParameterDeclaration: CustomStringConvertible {
                 guard let parameter = t.params.first,
                       case .parameter(let p) = parameter,
                       case .stringLiteral(let key) = p,
-                      key.count > 0,
                       let export = imports[key]?.body.first,
                       case .expression(let exp) = export,
                       exp.count == 1,
--- a/Sources/LeafKit/LeafLexer/LeafToken.swift
+++ b/Sources/LeafKit/LeafLexer/LeafToken.swift
@@ -181,9 +181,9 @@ public indirect enum ParameterDeclaration: CustomStringConvertible {
 
     public var description: String {
         switch self {
-            case .parameter(let p):  return p.description
-            case .expression(let p): return name + "(\(p.describe()))"
-            case .tag(let t):        return "tag(\(t.name): \(t.params.describe(",")))"
+            case .parameter(let p): return p.description
+            case .expression(_):    return self.short
+            case .tag(let t):       return "tag(\(t.name): \(t.params.describe(",")))"
         }
     }
 
@@ -206,8 +206,14 @@ public indirect enum ParameterDeclaration: CustomStringConvertible {
     func imports() -> Set<String> {
         switch self {
             case .parameter(_): return .init()
-            case .tag(let t): return t.imports()
             case .expression(let e): return e.imports()
+            case .tag(let t):
+                guard t.name == "import" else { return t.imports() }
+                guard let parameter = t.params.first,
+                      case .parameter(let p) = parameter,
+                      case .stringLiteral(let key) = p,
+                      key.count > 0 else { return .init() }
+                return .init(arrayLiteral: key)
         }
     }
     
@@ -215,12 +221,18 @@ public indirect enum ParameterDeclaration: CustomStringConvertible {
         switch self {
             case .parameter(_): return self
             case .tag(let t):
-                guard !t.imports().isEmpty else { return self }
-                let new = t.inlineRefs([:], imports).first!
-                switch new {
-                    case .custom(let t): return .tag(t)
-                    default: fatalError("unreachable")
+                guard t.name == "import" else {
+                    return .tag(.init(name: t.name, params: t.params.inlineImports(imports)))
                 }
+                guard let parameter = t.params.first,
+                      case .parameter(let p) = parameter,
+                      case .stringLiteral(let key) = p,
+                      key.count > 0,
+                      let export = imports[key]?.body.first,
+                      case .expression(let exp) = export,
+                      exp.count == 1,
+                      let e = exp.first else { return self }
+                return e                    
             case .expression(let e):
                 guard !e.isEmpty else { return self }
                 return .expression(e.inlineImports(imports))
@@ -333,19 +345,27 @@ internal extension Array where Element == ParameterDeclaration {
     
     func inlineImports(_ imports: [String : Syntax.Export]) -> [ParameterDeclaration] {
         guard !self.isEmpty else { return self }
-        var imports = imports
-        imports = imports.filter {
-            // Any imports that aren't purely a single raw/variable are non-sensical
-            // as a parameter value, so strip those out
-            if $0.value.body.count != 1 { return false }
-            switch $0.value.body.first {
-                case .variable(_),
-                     .raw(_): return true
-                default: return false
-            }
-        }
         guard !imports.isEmpty else { return self }
         return self.map { $0.inlineImports(imports) }
+    }
+    
+    func atomicRaw() -> Syntax? {
+        // only atomic expressions can be converted
+        guard self.count < 2 else { return nil }
+        var buffer = ByteBufferAllocator().buffer(capacity: 0)
+        // empty expressions = empty raw
+        guard self.count == 1 else { return .raw(buffer) }
+        // only single value parameters can be converted
+        guard case .parameter(let p) = self[0] else { return nil }
+        switch p {
+            case .constant(let c): buffer.writeString(c.description)
+            case .keyword(let k): buffer.writeString(k.rawValue)
+            case .operator(let o): buffer.writeString(o.rawValue)
+            case .stringLiteral(let s): buffer.writeString(s)
+            // .tag, .variable not atomic
+            default: return nil
+        }
+        return .raw(buffer)
     }
 }
 // MARK: --- END OF SECTION TO BE MOVED ---
--- a/Sources/LeafKit/LeafLexer/LeafToken.swift
+++ b/Sources/LeafKit/LeafLexer/LeafToken.swift
@@ -202,6 +202,30 @@ public indirect enum ParameterDeclaration: CustomStringConvertible {
             case .tag:        return "tag"
         }
     }
+    
+    func imports() -> Set<String> {
+        switch self {
+            case .parameter(_): return .init()
+            case .tag(let t): return t.imports()
+            case .expression(let e): return e.imports()
+        }
+    }
+    
+    func inlineImports(_ imports: [String : Syntax.Export]) -> ParameterDeclaration {
+        switch self {
+            case .parameter(_): return self
+            case .tag(let t):
+                guard !t.imports().isEmpty else { return self }
+                let new = t.inlineRefs([:], imports).first!
+                switch new {
+                    case .custom(let t): return .tag(t)
+                    default: fatalError("unreachable")
+                }
+            case .expression(let e):
+                guard !e.isEmpty else { return self }
+                return .expression(e.inlineImports(imports))
+        }
+    }
 }
 
 internal extension Array where Element == ParameterDeclaration {
@@ -299,6 +323,29 @@ internal extension Array where Element == ParameterDeclaration {
     
     func describe(_ joinBy: String = " ") -> String {
         return self.map {$0.short }.joined(separator: joinBy)
+    }
+    
+    func imports() -> Set<String> {
+        var result = Set<String>()
+        self.forEach { result.formUnion($0.imports()) }
+        return result
+    }
+    
+    func inlineImports(_ imports: [String : Syntax.Export]) -> [ParameterDeclaration] {
+        guard !self.isEmpty else { return self }
+        var imports = imports
+        imports = imports.filter {
+            // Any imports that aren't purely a single raw/variable are non-sensical
+            // as a parameter value, so strip those out
+            if $0.value.body.count != 1 { return false }
+            switch $0.value.body.first {
+                case .variable(_),
+                     .raw(_): return true
+                default: return false
+            }
+        }
+        guard !imports.isEmpty else { return self }
+        return self.map { $0.inlineImports(imports) }
     }
 }
 // MARK: --- END OF SECTION TO BE MOVED ---
--- a/Sources/LeafKit/LeafParser/LeafParser.swift
+++ b/Sources/LeafKit/LeafParser/LeafParser.swift
@@ -186,18 +186,31 @@ struct LeafParser {
         }
 
         if case .conditional(let new) = newSyntax {
-            switch new.condition {
-            // a new if, never attaches to a previous
-            case .if:
-                append(newSyntax)
-            case .elseif, .else:
-                // elseif and else ALWAYS attach
-                // ensure there is a leading conditional to
-                // attach to, prioritize waiting bodies, then check finished
-                guard let last = awaitingBody.last?.body.last ?? finished.last, case .conditional(let tail) = last else {
-                    throw "unable to attach \(new.condition) to \(finished.last?.description ?? "<>")"
-                }
-                try tail.attach(new)
+            guard let conditional = new.chain.first else { throw "Malformed syntax block" }
+            switch conditional.0.naturalType {
+                // a new if, never attaches to a previous
+                case .if:
+                    append(newSyntax)
+                case .elseif, .else:
+                    let aW = awaitingBody.last?.body
+                    let previousBlock: Syntax?
+                    switch aW {
+                        case .none: previousBlock = finished.last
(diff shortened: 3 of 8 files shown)


Sources/LeafKit/LeafAST.swift
```diff
@@ -44,21 +44,30 @@ public struct LeafAST: Hashable {
...
-                        pos = ast.index(after: pos)
-                    }
-                default:
-                    var new: Syntax? = nil
-                    unresolvedRefs.formUnion(ast[pos].inlineRefs(externals, &new))
-                    if let new = new { ast[pos] = new }
-                    pos = ast.index(after: pos)
+            // get desired externals for this Syntax - if none, continue
+            let wantedExts = ast[pos].externals()
+            if wantedExts.isEmpty {
```
tdotclare, line 49: Top level AST resolution no longer requires variant behavior from different Syntaxes - a Syntax now simply reports what externals it wants at any point in the nested tree, and when called to resolve itself with a list of external files (and when recursively called from an Extend, applicable imports) and then returns an updated [Syntax] list.

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -68,28 +172,110 @@ extension Syntax {
...
+        }
+        func imports() -> Set<String> {
+            return importSet
+        }
+        
+        func inlineRefs(_ externals: [String: LeafAST], _ imports: [String : Syntax.Export]) -> [Syntax] {
+            var newExports = [String: Export]()
+            var newImports = imports
+            var newExternalsSet = Set<String>()
+            var newImportSet = Set<String>()
```
tdotclare, line 225: Syntax.Extend.inlineRefs is the main behavioral interface for extending; at the top level of a document, no imports exist yet. Once an Extend is hit, the Extend will then inject any external references into it's own Exports, rebuilding those first, and then loading the external template it references if available, directly rebuilding the external Syntax as it assembles the current resolvable state of the Syntax - then passing the complete [Syntax] back to the top level of the AST (or to the higher level of the AST tree if nested within another Syntax).

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -135,24 +344,73 @@ extension Syntax {
...
 
-    public final class Conditional {
-        public let condition: ConditionalSyntax
-        public internal(set) var body: [Syntax]
-        public private(set) var next: Conditional?
+    public struct Conditional: BodiedSyntax {
+        public internal(set) var chain: [(
+            condition: ConditionalSyntax,
+            body: [Syntax]
+        )]
```
tdotclare, line 351: Syntax.Conditional is now a struct containing a chain of (ConditionalSyntax, Body) block instead of a set of linked class instances.

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -285,111 +597,3 @@ extension Syntax: CustomStringConvertible {
         }
     }
 }
-
-extension Syntax {
-    mutating func inlineRefs(_ externals: [String: LeafAST], _ new: inout Syntax?) -> Set<String> {
-        var changed = false
-        var rf = Set<String>()
```
tdotclare, line 292: Monolithic inlining function obviated by BodiedSyntax directed resolution.

Tests/LeafKitTests/LeafKitTests.swift
```diff
@@ -34,7 +34,7 @@ final class ParserTests: XCTestCase {
 
         let expectation = """
```
tdotclare, line 35: Various minor internal changes to print formatting of [ParameterDeclaration] for clarity

Tests/LeafKitTests/LeafKitTests.swift
```diff
@@ -1101,6 +1101,119 @@ final class LeafKitTests: XCTestCase {
             XCTAssertEqual(page.string, expected)
     }
 
+    func testGH57_LoopedConditionalImport() throws {
+        var test = TestFiles()
```
tdotclare, line 1105: Issue #57 tests

Tests/LeafKitTests/LeafKitTests.swift
```diff
@@ -1101,6 +1101,119 @@ final class LeafKitTests: XCTestCase {
...
+            "list2": ["A", "B", "C"],
+        ]).wait()
+        XCTAssertEqual(page.string, expected)
+    }
+    
+    func testImportParameter() throws {
+        var test = TestFiles()
+        test.files["/base.leaf"] = """
+        #extend("parameter"):
+            #export("admin", admin)
```
tdotclare, line 1185: Test context variable as #export value

Tests/LeafKitTests/LeafKitTests.swift
```diff
@@ -1101,6 +1101,119 @@ final class LeafKitTests: XCTestCase {
...
+        #endextend
+        """
+        test.files["/delegate.leaf"] = """
+        #extend("parameter"):
+            #export("delegated", false || bypass)
```
tdotclare, line 1190: Test expression with keyword and context variable as #export value

Tests/LeafKitTests/LeafTests.swift
```diff
@@ -516,9 +516,9 @@ final class LeafTests: XCTestCase {
         """
 
         let syntax = """
-        [variable(index), operator(-), constant(5)]
-        [constant(10), operator(-), constant(5)]
-        [constant(10), operator(-), constant(5)]
+        expression[variable(index), operator(-), constant(5)]
```
tdotclare, line 519: Minor .expression format changes

tdotclare:
A future PR could/should enable Swift-style optional tests as well to allow providing defaults where no value is exported - EG:
```
#import("baseURL" ?? "www.default.com")
```

ffried:
@tdotclare A quick thought, that came to me today:
Where are the exported expessions evaluated? At import-time or at export-time?

Given the following:

```
#export("test", import("somethingElse"))
#export("somethingElse", "abc")
```

And then:
```
<div>#import("test")</div>
```

Will this try to import "somethingElse" inside the context where "test" is exported? Or will it (recursively) import it when "test" is imported?

The latter would make sense to me, allowing for quite flexible patterns. The former is probably easier to implement, I guess... 🤔

tdotclare:
It depends; if the export value is a single resolvable value during AST handling (eg, `#export("key", "value")` - with value being a string literal, a numeric constant, or a keyword, then a top-level #import syntax will automatically resolve to raw data, so yes, the AST would automatically become a single raw data. If the value is an actual expression like `#export("key", variable == true)` then the imported expression won't be resolved until render time when the variable exists.

Sources/LeafKit/LeafSerialize/LeafSerializer.swift
```diff
@@ -61,24 +61,15 @@ struct LeafSerializer {
...
-            $0.bool ?? !$0.isNull
-        }.reduce(true) { $0 && $1 }
-        if satisfied {
-            try serialize(body: conditional.body)
-        } else if let next = conditional.next {
-            try serialize(next)
+        evaluate:
+        for block in conditional.chain {
+            let expression = try self.resolve(parameters: block.condition.expression())
+            let satisfied = expression.map { $0.bool ?? !$0.isNull }
```
ffried, line 67: Using `allSatisfy`, this won't have to iterate over all elements. The first `false` expression will result in `false` either way.

Sources/LeafKit/LeafLexer/LeafToken.swift
```diff
@@ -300,6 +336,37 @@ internal extension Array where Element == ParameterDeclaration {
     func describe(_ joinBy: String = " ") -> String {
         return self.map {$0.short }.joined(separator: joinBy)
     }
+    
+    func imports() -> Set<String> {
+        var result = Set<String>()
+        self.forEach { result.formUnion($0.imports()) }
```
ffried, line 342: NitPick: `return reduce(Set<String>()) { $0.formUnion($1.imports()) }`
Or even: `Set(lazy.flatMap { $0.imports() })`

Sources/LeafKit/LeafLexer/LeafToken.swift
```diff
@@ -202,6 +202,42 @@ public indirect enum ParameterDeclaration: CustomStringConvertible {
...
+        switch self {
+            case .parameter(_): return self
+            case .tag(let t):
+                guard t.name == "import" else {
+                    return .tag(.init(name: t.name, params: t.params.inlineImports(imports)))
+                }
+                guard let parameter = t.params.first,
+                      case .parameter(let p) = parameter,
+                      case .stringLiteral(let key) = p,
+                      key.count > 0,
```
ffried, line 230: NitPick: `!key.isEmpty`

Sources/LeafKit/LeafLexer/LeafToken.swift
```diff
@@ -202,6 +202,42 @@ public indirect enum ParameterDeclaration: CustomStringConvertible {
...
+    func imports() -> Set<String> {
+        switch self {
+            case .parameter(_): return .init()
+            case .expression(let e): return e.imports()
+            case .tag(let t):
+                guard t.name == "import" else { return t.imports() }
+                guard let parameter = t.params.first,
+                      case .parameter(let p) = parameter,
+                      case .stringLiteral(let key) = p,
+                      key.count > 0 else { return .init() }
```
ffried, line 215: NitPick: `!key.isEmpty`

Sources/LeafKit/LeafParser/LeafParser.swift
```diff
@@ -186,18 +186,31 @@ struct LeafParser {
...
+                    }
+                    guard let existingConditional = previousBlock,
+                        case .conditional(var tail) = existingConditional else {
+                            throw "Can't attach \(conditional.0) to \(previousBlock?.description ?? "empty AST")"
+                    }
+                    try tail.attach(new)
+                    switch aW {
+                        case .none:
+                            finished.removeLast()
+                            finished.append(.conditional(tail))
```
ffried, line 209: `finished[finished.index(before: finished.endIndex)] = .conditional(tail)` wouldn't have to modify the size of `finished`.

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -18,29 +18,133 @@ public enum ConditionalSyntax {
     case `if`([ParameterDeclaration])
     case `elseif`([ParameterDeclaration])
     case `else`
+    
+    func imports() -> Set<String> {
+        switch self {
+            case .if(let pDA), .elseif(let pDA):
+                var imports = Set<String>()
```
ffried, line 25: Similar to above, this could be one of:
- `return pDA.reduce(Set<String>()) { $0.formUnion($1.imports()) }`
- `return Set(pDA.lazy.flatMap { $0.imports() })`

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -18,29 +18,133 @@ public enum ConditionalSyntax {
...
+internal protocol BodiedSyntax {
+    func externals() -> Set<String>
+    func imports() -> Set<String>
+    func inlineRefs(_ externals: [String: LeafAST], _ imports: [String: Syntax.Export]) -> [Syntax]
+}
+
+extension Array: BodiedSyntax where Element == Syntax {
+    func externals() -> Set<String> {
+        var result = Set<String>()
+        _ = self.map { result.formUnion( $0.externals()) }
```
ffried, line 134: Similar to above, this could be one of:
- `return reduce(Set<String>()) { $0.formUnion($1.externals()) }`
- `return Set(lazy.flatMap { $0.externals() })`

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -18,29 +18,133 @@ public enum ConditionalSyntax {
...
+        return result
+    }
+    
+    func imports() -> Set<String> {
+        var result = Set<String>()
+        _ = self.map { result.formUnion( $0.imports() ) }
```
ffried, line 140: Similar to above, this could be one of:
- `return reduce(Set<String>()) { $0.formUnion($1.imports()) }`
- `return Set(lazy.flatMap { $0.imports() })`

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -18,29 +18,133 @@ public enum ConditionalSyntax {
...
+        return result
+    }
+
+    func inlineRefs(_ externals: [String: LeafAST], _ imports: [String: Syntax.Export]) -> [Syntax] {
+        var result = [Syntax]()
+        _ = self.map { result.append(contentsOf: $0.inlineRefs(externals, imports)) }
```
ffried, line 146: This is probably easier expressed as: `return flatMap { $0.inlineRefs(externals, imports) }`

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -68,28 +172,110 @@ extension Syntax {
...
         public private(set) var exports: [String: Export]
+        private var externalsSet: Set<String>
+        private var importSet: Set<String>
 
         public init(_ params: [ParameterDeclaration], body: [Syntax]) throws {
             guard params.count == 1 else { throw "extend only supports single param \(params)" }
             guard case .parameter(let p) = params[0] else { throw "extend expected parameter type, got \(params[0])" }
             guard case .stringLiteral(let s) = p else { throw "import only supports string literals" }
             self.key = s
+            self.externalsSet = .init(arrayLiteral: self.key)
```
ffried, line 186: NitPick: `self.externalsSet = [self.key]`

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -68,28 +172,110 @@ extension Syntax {
...
+                    continue
+                }
+                guard case .export(let e) = value.inlineRefs(externals, imports).first else { fatalError() }
+                newExports[key] = e
+                newExternalsSet.formUnion(e.externals())
+                newImportSet.formUnion(e.imports())
+            }
+            
+            // Now add this extend's exports onto the passed imports
+            newExports.forEach {
```
ffried, line 240: This would allow `Dictionary` to resize and merge values in one go: `newExports.merge(newImports, uniquingKeysWith: { $1 })`

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -68,28 +172,110 @@ extension Syntax {
...
+                // Remove this AST from the externals to avoid needless checks
+                let externals = externals.filter { $0.key != self.key }
+                newAst.ast.forEach {
+                    // Add each external syntax, resolving with the current available
+                    // exports and passing this extend's exports to the syntax's imports
+                    
+                    results += $0.inlineRefs(externals, newImports)
+                    // expressions may have been created by imports, convert
+                    // single parameter static values to .raw
+                    if case .expression(let e) = results.last {
```
ffried, line 265: The nested ifs could be combined: `if case .expression(let e) = results.last, let raw = e.atomicRaw() { ... }`

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -165,37 +423,39 @@ extension Syntax {
...
     }
 
-    public struct Loop {
+    public struct Loop: BodiedSyntax {
         /// the key to use when accessing items
         public let item: String
         /// the key to use to access the array
         public let array: String
 
         /// the body of the looop
```
ffried, line 454: Typo: 'looop' 🙂

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
```diff
@@ -248,10 +531,39 @@ extension Syntax {
...
+        }
+        
+        func imports() -> Set<String> {
+            return importSet
+        }
+        
+        func inlineRefs(_ externals: [String: LeafAST], _ imports: [String : Syntax.Export]) -> [Syntax] {
+            guard !importSet.isEmpty || !externalsSet.isEmpty else { return [.custom(self)] }
+            let p = params.imports().isEmpty ? params : params.inlineImports(imports)
+            let b = body == nil ? nil : body!.inlineRefs(externals, imports)
```
ffried, line 564: Could simply be: `let b = body?.inlineRefs(externals, imports)`

ffried:
> It depends; if the export value is a single resolvable value during AST handling (eg, `#export("key", "value")` - with value being a string literal, a numeric constant, or a keyword, then a top-level #import syntax will automatically resolve to raw data, so yes, the AST would automatically become a single raw data. If the value is an actual expression like `#export("key", variable == true)` then the imported expression won't be resolved until render time when the variable exists.
> 
> * Note - in the case where the expression is not resolvable at parsing time (like `variable == true`) - the import is still fully resolved during parsing, removing the `import` statement - it is replaced by the actual expression. EG; if written out, `#import("key")` would become `#(variable == true)` in the AST.

Wow, cool. That sounds like what I had hoped for! Thanks again for this!

Sources/LeafKit/LeafSerialize/LeafSerializer.swift
tdotclare: Looked at this again, it's actually simpler - the parameter resolution inherently should resolve to a single value'd LeafData when properly structured so it's turning into this, where `resolveAtomic()` is a new internal function of Serializer which guards that a parameter list has exactly one parameter before trying to resolve, and returns the resolved single value or null if the parameter is un-resolvable
```swift
let evaluated = try self.resolveAtomic(parameters: block.condition.expression())
if evaluated.bool ?? !evaluated.isNull {
  try serialize(body: block.body)
  break evaluate
}
```

ffried: Ha, even better! 👍

 Small NitPick: the `count == 0` check in `resolveAtomic` could be replaced by `isEmpty`. 😋

Sources/LeafKit/LeafLexer/LeafToken.swift
tdotclare: Removed since the `let export = imports[key]?...` would catch an empty set anyway

Sources/LeafKit/LeafLexer/LeafToken.swift
tdotclare: Changed - I don't know why I reflexively avoid `isEmpty`, hah

Sources/LeafKit/LeafLexer/LeafToken.swift
tdotclare: Leaving as-is for clarity; `reduce` won't compile as you wrote it and while I'd originally tried to use it anyway, it gets too wonky looking IMO. Just not a fan of lazy in this usage since it masks what's happening

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
tdotclare: Leaving as-is for clarity - side note btw - given that these functions are only called during template resolution (never during data serialization), I'm not super concerned about any minor performance differences and prefer clear behavior :)

Sources/LeafKit/LeafLexer/LeafToken.swift
ffried: Ah, should have been `reduce(into: Set<String>()) { ... }`, sorry. The `into:` version passes a mutating parameter into the closure.

Same goes for the other similar cases I mentioned.

Sources/LeafKit/LeafParser/LeafParser.swift
tdotclare: 👍 think I'd just c/p'd that from append() and my eyes had glazed over at that point

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
ffried: As above, the `into:` label is missing in my code. But as long as it’s not performance critical, leaving it as is is also fine. 🙂

tdotclare: IIRC I tried it with reduce originally but could swear I got compiler issues with it having trouble on type inference, and it ended up being even more awkward than the foreach version

Sources/LeafKit/LeafSerialize/LeafSerializer.swift
```diff
@@ -172,6 +161,18 @@ struct LeafSerializer {
         )
         return try resolver.resolve().map { $0.result }
     }
+    
+    // Directive resolver for a [ParameterDeclaration] where only one parameter is allowed that must resolve to a single value
+    private func resolveAtomic(parameters: [ParameterDeclaration]) throws -> LeafData {
```
tdotclare, line 166: New internal resolver for fixed-pattern `[ParameterDeclaration]` where one and exactly one Parameter is required, returning the resolved atomic LeafData or null if un-resolvable and throwing if count is wrong.

Sources/LeafKit/LeafSyntax/LeafSyntax.swift
ffried: As long as you qualify the Element Parameter of Set, the closure should be typed enough to infer all remaining types. But yea, sometimes weird things happen with type inference. 😀

tdotclare: Yeah - should clarify, `reduce(into` was fine for *compiling* but was incapable of inferring $0 and $1 in the closure, and I found that obnoxious :)

ffried: @tdotclare Hehe, the "while you're typing" problem is because there are two parameters in the closure ($0 and $1). As long as you only use one, the compiler is unsure whether you're going to refer to the second one as `$0.1` or `$1`.
For getting auto-completion while typing, what works is this:
```swift
return pDA.reduce(into: Set<String>()) {
    let imports = $1.imports()
    $0.formUnion(imports)
}
```

I know it's dumb. They should actually ban the `$0.1` access to non-tuple parameters in closures in which case the compiler would know the type of `$0` while you're typing.

PR #62, opened 2020-06-26
Add convenience methods to obscure some of the inner NIO workings

MPLew-is:
In usage outside of Vapor, these methods will provide a much cleaner API and eliminate a lot of the boilerplate code needed to manage the NIO objects. Most of the implementation comes from the existing tests, this change mostly just provides that common usage pattern as a dedicated initializer.

Not sure if this is wanted, but figured I'd get an implementation out to see. This would certainly remove quite a bit of boilerplate code from a few simple projects I've been doing, and could be used down the road to simplify some of the existing tests as well, which is where this code was pulled from.

### DIFF
--- a/Sources/LeafKit/LeafRenderer.swift
+++ b/Sources/LeafKit/LeafRenderer.swift
@@ -74,7 +74,70 @@ public final class LeafRenderer {
         self.eventLoop = eventLoop
         self.userInfo = userInfo
     }
-    
+
+
+    /// An optional thread pool, initialized only in the convenience initializer
+    /// `init(configuration:tags:cache:userInfo:)` to provide an initializer with all parameters
+    /// having default values
+    private var threadPool: NIOThreadPool?
+
+    /// Whether `eventLoop` should be shut down on `deinit`, only set to `true` in the convenience
+    /// initializer `init(configuration:tags:cache:userInfo:)` to provide an initializer with all
+    /// parameters having default values
+    private var shouldShutDownEventLoop = false
+
+    /// Initializes a `LeafRenderer` with defaults for all parameters and self-created NIO objects.
+    ///
+    /// This initializer is designed to facilitate simple use of leaf templates outside the context
+    /// of Vapor (or other applications making heavy use of NIO) by managing the initialization and
+    /// destruction of any needed NIO primitives, as well as setting some additional default
+    /// configuration values.
+    ///
+    /// All overlapping parameters (`tags`, `cache`, and `userInfo`) have the same meaning and
+    /// default values as in `init(configuration:tags:cache:sources:eventLoop:userInfo:)`, as this
+    /// is intended to be as close to a pass-through initializer as possible.
+    ///
+    /// - Parameters:
+    ///   - rootDirectory: root directory for configurations, defaulting to the working directory
+    ///   - tags:          dictionary of tags, defaulting to the normal `defaultTags` property
+    ///   - cache:         cache object, defaulting to an instance of `DefaultLeafCache`
+    ///   - userInfo:      custom instance data, defaulting to an empty dictionary
+    public convenience init(
+        rootDirectory: String = "./",
+        tags: [String: LeafTag] = defaultTags,
+        cache: LeafCache = DefaultLeafCache(),
+        userInfo: [AnyHashable: Any] = [:]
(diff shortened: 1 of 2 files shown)


tdotclare:
Couple issues with this unfortunately:

* LeafRenderer is not intended to be a fully encapsulated object for the purposes of operating by itself -
  * In practice, when bound to Vapor, an instance of LeafRenderer is created on every request and is provided the shared objects. As a result, if you attempt to encapsulate the NIO objects into one instance, that means significant overhead and more importantly - a lack of combined resources (eg, caching will not be persisted).
  * Note that all stored objects to LeafRenderer are classes, not structs - this is intentional to persist data by having shared, thread-safe single instances that multiple LeafRenderer instances can use. While it's *possible* to use embeddedEventLoops as a convenience for a persisted, single-threaded renderer if only one is needed and you will only ever render one template at a time, it's still very unperformant and should be avoided.

Adding such convenience initializers for public use would be a bad idea IMO.

However - I assume since you're looking for something like what I just described (a single global instance when using LeafKit by itself) - you may find it simpler just to create your own wrapper. LeafRenderer really only has one public interface currently - `render()`.

I'd suggest looking at `TestRenderer` in [TestHelpers.swift](https://github.com/vapor/leaf-kit/blob/27724d049e959bef6bfbf71be33b446fccfd5e3f/Tests/LeafKitTests/TestHelpers.swift#L49) for an example of a basic convenience object which simply wraps a LeafRenderer and passes through to the actual object when called.

MPLew-is:
It's perfectly fine if this isn't desired, but I guess I'm still not fully understanding the reasoning behind that. The implementation in this PR is a purely additive change that I've been careful to make sure would not affect the existing usage of `LeafRenderer` at all; it just provides some convenience methods to facilitate easier use outside Vapor, so the three points above are not directly applicable in my view.

Mostly, a fair amount of knowledge about NIO is required to get rendering in Leaf off the ground in a simple command-line project for instance, which to me feels directly contradictory to the general Swift principle of progressive disclosure and causes a pretty large impediment to further adoption of this templating language (which is one of the better ones I've come across). However, if adoption outside of Vapor isn't a goal of this project that's completely fine and I'll revert back to just using the wrapper I made from which this code was derived; I would however suggest making that clear in the documentation, that this is only intended for use in Vapor.

tdotclare:
Hi Mike, apologies if I sounded terse - 

I agree that an encapsulated handler would be great to have at some point, but unless it can be guaranteed to operate properly in isolation (and LeafKit isn't at that point yet) - public methods doing so aren't something I'd want to put in yet, and they'd better be addressed by making a separate class (like `LeafEngine` or something) that explicitly is intended to be a singleton object.

Note that your PR failed on thread safety - setting up NIO objects has to be heavily double-checked to confirm that the pools are properly being set up and dismantled.

It's something I'd love to formalize more when I've finished rebuilding the syntax to a point that safety is guaranteed through the pipeline, and that needs to be done before discrete engine support as an object is feasible.

MPLew-is:
Yeah, this PR was mostly a proposal for a starting point; I fully realize that it would need some changes, including fixing the thread sanitizer failures. I’m happy to put in some of the work to get this feature more stable than this initial PR but only if it’s something that is or will be desired, even if it won’t be fully merged for a while to come pending the other changes you talked about.

Just my two cents, but I also think it’ll present an easier public interface and better progressive disclosure if everything is integrated into `LeafRenderer` instead of split out into another class; this allows a user to gradually incorporate more and more NIO features as they’re needed in their project and have them coexist side-by-side with more synchronous methods without having to refactor their codebase to do so.

tdotclare:
Yeah - I 100% understand the motivation, just would not suggest attempting to stabilize this PR anytime soon because too much is changing internally, and I'm actively trying to lock down *any* public changes at all until internal changes are stable enough to have consistent interfaces to `LeafRenderer` - this is more of a "it's not worth doing anything about it yet" comment than anything else.

tdotclare:
* To clarify - part of why I don't want anything public like that yet is that because *the templating language* is still having features added, and some holdover from Leaf3 is currently being kept purely for consistency with Vapor - I don't want those behaviors to necessarily be part of a "public, non-Vapor" usage and would prefer not to have ingrained stuff hanging around from the transition

PR #70, opened 2020-09-14
Make LeafAST.name public

123FLO321:
- make `name` property on `LeafAST` public

### DIFF
--- a/Sources/LeafKit/LeafAST.swift
+++ b/Sources/LeafKit/LeafAST.swift
@@ -5,12 +5,13 @@
 /// `LeafAST` represents a "compiled," grammatically valid Leaf template (which may or may not be fully resolvable or erroring)
 public struct LeafAST: Hashable {
     // MARK: - Public
-    
+
+    public let name: String
+
     public func hash(into hasher: inout Hasher) { hasher.combine(name) }
     public static func == (lhs: LeafAST, rhs: LeafAST) -> Bool { lhs.name == rhs.name }
 
     // MARK: - Internal/Private Only
-    let name: String
 
     init(name: String, ast: [Syntax]) {
         self.name = name


123FLO321:
While implemented a custom `LeafCache`, I noticed that `insert` passes an instance of `LeafCache` while `retrieve` and `remove` use just the name.
And because the `name` property is `internal` it is impossible to implement a custom cache.

tdotclare:
Correct - this is an error which has not been patched due to below:

Next release refactors LeafCache's protocol interface to remove `name` as a caching key in favor of an opaque hashable struct (`LeafAST.Key`) for keying.

123FLO321:
@tdotclare That is just one example where I wanted to use it.
But in general it would probably not hurt to make the name public.

tdotclare:
There *is* a public computed property in next release for `LeafAST.name` - it's simply blocked at the moment to prevent adherence until the API changes are complete, and to avoid the implication that it is a unique reference (with multiple sources for templates, the name may collide).

123FLO321:
@tdotclare Makes sense.
In this case I'll close this issue and wait for the next release.

tdotclare:
If you're able to comment anything about your use case for a custom cache implementation btw, please let me know - may be something addressed in the next release already (like publicizing/accessing information about the parsed template structure)

123FLO321:
@tdotclare It's a very specific use-case.
I use Leaf to render JSON-templates that might change life in production (but they rarely do, related: https://github.com/vapor/leaf/pull/163).

My custom cache simply uses the default leaf cache in the background, but it clears specific keys if that file changes on disc.

tdotclare:
Gotcha. Yes - that's not something that's addressed yet due to overhead for monitoring changes, but next release does publicize timestamps for cached ASTs in order to facilitate that type of behavior in the future (eg auto-purging compiled ASTs when on-disk template is newer than the current cached version).

Additionally, next release introduces raw inlining behavior - eg, for cases like that where the input file is not parsed at all by Leaf but is simply inserted as raw data, and that is entirely uncached.

tdotclare:
Also RE: your linked issue on req.leaf; next version will have req.leaf.renderer public once I finish securing some things for the updates to LeafKit, but you can also generally use app.leaf.renderer to update the cache from requests as the cache is a shared object.

123FLO321:
@tdotclare thank you

On requests using `request.application.leaf.renderer.render()` isn't that easy because it's not running in the same event loop.
Simply calling `request.leaf.render()` is much more convenient since you don't need to worry about that.

Issue #69, opened 2020-07-27
Leaf-kit 1.0.0-rc.1.17 no longer compiling on raspberry pi


bottlehall:
Compile generates an error due to line 45 of file LeafKit/LeafData/LeafDataRepresentable.swift. The offending line is:
```swift
extension Float80: LeafDataRepresentable {}
```

Enclosing it in compiler directives fixes the problem:
```swift
#if arch(i386) || arch(x86_64)
extension Float80: LeafDataRepresentable {}
#endif
```

tdotclare:
Will patch in next release.

0xTim:
This also occurs on Apple Silicon

tdotclare:
Closed in LeafKit-1.0.0tau

PR #71, opened 2020-09-20
Fix compiler error on non-x86 architecture builds

bottlehall:
To fix compiler error on non-X86 architectures. Issue https://github.com/vapor/leaf-kit/issues/69 refers.

Adds compiler directive in LeafDataRepresentative.swift to remove redundant extension to Float80.


tdotclare:
Obviated by tau-1

PR #79, opened 2020-11-09
Leaf 4 Release

### LINKED ISSUE #78: #count tag not working
When trying to use `#count` tag in an `#if` condition `else` branch always gets executed whether the condition is satisfied or not. Also, when trying to display number of elements somewhere on the page, nothing renders. For example `<p>Total number of articles is #count(articles)</p>` renders `Total number of articles is`.

0xTim:
Final release for Leaf 4. You can trick the rendering engine into making comments but we'll leave it out as a secret feature for now until it's properly implemented. The release forgoes inline comments and auto-escaping of `#` for now, those can be added at a later date

Also resolves #78 and #80


Tests/LeafKitTests/GHTests/VaporLeaf.swift
```diff
@@ -94,21 +89,23 @@ final class GHLeafIssuesTestArchive: XCTestCase {
             try XCTAssertEqual(render(template, ["a": "a"]), expected)
         }
     }
-    
-    // TODO: This commenting style is not used anymore but needs a replacement
-    func _testGH127SingleLine() throws {
+
+    func testGH127SingleLine() throws {
```
siemensikkema, line 93: already tested in `LeafKitTests.testCommentSugar` and this test doesn't reflect the original issue anymore: https://github.com/vapor/leaf/issues/127

0xTim: @siemensikkema you're right, I've removed this

siemensikkema (approved):

PR #91, opened 2021-03-19
Correctly parsing non Leaf `#` tag.

Hailong:
It includes the following changes:

* Literally output the non Leaf `#` tags.
* Add the explicit support for `#comment` tag.
* Enable the corresponding unit test.

### DIFF
--- a/Sources/LeafKit/LeafSerialize/LeafSerializer.swift
+++ b/Sources/LeafKit/LeafSerialize/LeafSerializer.swift
@@ -76,7 +76,7 @@ internal struct LeafSerializer {
             body: tag.body,
             userInfo: self.userInfo
         )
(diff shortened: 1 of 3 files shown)


0xTim:
@Hailong thanks for this, honestly not sure how it was such a simple fix 😅 

Could you change https://github.com/vapor/leaf-kit/blob/main/.github/workflows/test.yml#L24 to point to `main` instead of `master` - that should get CI working

0xTim (approved):

0xTim:
Original comment:

It includes the following changes:

* Literally output the non Leaf `#` tags.
* Add the explicit support for `#comment` tag.
* Enable the corresponding unit test.

0xTim:
@Hailong never mind about the fix, I can edit your branch

PR #95, opened 2021-11-18
Ignore unfound import

### LINKED ISSUE #84: Allow `#import` to be empty or not declared
**Is your feature request related to a problem? Please describe.**
I have a `layout.leaf` looking like this

```html
<doctype html>
<html>
    <head>
        <link rel="stylesheet" href="mystyles.css"/>
        #import("otherStyles")
    </head
    <body>
        #import("content")
    </body>
</html>
```

But if I don't export it, I'm having the following error:

> "import(\"styles\") should have been resolved BEFORE serialization".

What I have to do now to have it working is adding `#export("otherStyles"): #endexport` to all of my views, even if they don't have otherStyles.

**Describe the solution you'd like**
I would like that import to be empty, I don't always want to export additional styles, possible solutions are

```
a) #import("otherStyles") // Modify default behaviour to be optional or allow empty imports
b) #import("otherStyles", allowEmpty: true)
c) #import("otherStyles", "default")
```

`a` or `c` would be perfect 👍

fananek:
This makes the renderer and serializer ignore imports in case no corresponding export has been found rather than erroring by default. This also adds a new configuration flag `ignoreUnfoundImports` to allow you to set the behaviour to error on non-existent imports

Fixes #84

### DIFF
--- a/Sources/LeafKit/LeafSerialize/LeafSerializer.swift
+++ b/Sources/LeafKit/LeafSerialize/LeafSerializer.swift
@@ -52,7 +52,7 @@ internal struct LeafSerializer {
         guard resolved.count == 1, let leafData = resolved.first else {
             throw "expressions should resolve to single value"
         }
-        try? leafData.serialize(buffer: &self.buffer)
+        try? leafData.htmlEscaped().serialize(buffer: &self.buffer)
     }
 
     private mutating func serialize(body: [Syntax]) throws {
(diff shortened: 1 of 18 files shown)


Sources/LeafKit/LeafSerialize/LeafSerializer.swift
```diff
@@ -42,8 +46,12 @@ internal struct LeafSerializer {
             case .conditional(let c):  try serialize(c)
             case .loop(let loop):      try serialize(loop)
             case .expression(let exp): try serialize(expression: exp)
-            case .import, .extend, .export:
+            case .extend, .export:
                 throw "\(syntax) should have been resolved BEFORE serialization"
+            case .import:
+                if (!self.ignoreUnfoundImports) {
+                    throw "\(syntax) should have been resolved BEFORE serialization"
+                }
```
0xTim, line 54: I think we can tidy this up a bit
```suggestion
            case .import:
                if (!self.ignoreUnfoundImports) {
                    fallthrough
                } else {
                    break
                }
            case .extend, .export:
                throw "\(syntax) should have been resolved BEFORE serialization"
```

0xTim (changes requested):
One nit but otherwise this is great! Thanks

Sources/LeafKit/LeafSerialize/LeafSerializer.swift
0xTim: PS this may not compile, if it doesn't we can stick with the original

fananek: Hi Tim, thanks for suggestion. It compiles fine. At the end I removed the negation from a condition and switched break and fallthrough so it's one operation less.
```swift
case .import:
  if (self.ignoreUnfoundImports) {
     break
  } else {
    fallthrough
  }
case .extend, .export:
  throw "\(syntax) should have been resolved BEFORE serialization"
```

0xTim (approved):
LGTM!

PR #96, opened 2022-01-02
WIP: Added async signature to LeafRenderer

bdrelling:
I'm wondering if it would be beneficial to add an `async` overload to the `render(path:context:)` method. I'm opening this PR with an example of what I'm referring to, and if it seems useful, I am happy to clean it up with documentation and any other changes requested.

I've done the same thing in my [PlotVapor](https://github.com/bdrelling/PlotVapor/blob/main/Source/PlotRenderer.swift#L33-L35) package, and the ability to simply change the closure signature without having to append `.get()` to every `render` call is very nice.

If this is seen as a negative addition to the SDK, I would love to understand why as well so I can make the same consideration in my related package. 

Thank you!

### DIFF
--- a/Sources/LeafKit/LeafRenderer.swift
+++ b/Sources/LeafKit/LeafRenderer.swift
@@ -80,6 +80,10 @@ public final class LeafRenderer {
         }
     }
     
+    public func render(path: String, context: [String: LeafData]) async throws -> ByteBuffer {
+        try await render(path: path, context: context).get()
+    }
+    
     
     // MARK: - Internal Only
     /// Temporary testing interface


0xTim:
@bdrelling the main reason this hasn't been done yet is that very few people will actually talk to LeafKit directly. 99% of the access will be through Vapor's `View` protocol, with the async variants defined [here](https://github.com/vapor/vapor/blob/main/Sources/Vapor/Concurrency/ViewRenderer%2BConcurrency.swift).

That said, we will flatten it all eventually so this is a great start. All that's left to do is the function documentation and PR title/description (this is used for the release notes). Ideally we'd add a test for the async function as well if you want to explore that as well

PR #114, opened 2022-08-18
Fix downstream tests

0xTim:
Fixes the downstream tests for vapor/leaf by setting the required environment variable


0xTim:
Merging because I can 😛

PR #111, opened 2022-08-13
Adds a new `#with()` tag to make it easier to embed and extend tags

pontaoski:
Adds a new `#with()` tag to make it easier to work with extend and embed tags:

```leaf
#with(parent):
  #(child)
#endwith
```

PR #115, opened 2022-08-20
Draft: some rewrite thing

pontaoski:
This PR is some sort of mess of me deciding to make the scanner have better errors, then ending up rewriting the scanner, then ending up rewriting the parser that consumes that scanner, then ending up rewriting the evaluators that consume ASTs from the parser.


Tests/LeafKitTests/GHTests/VaporLeaf.swift
```diff
@@ -50,21 +50,14 @@ final class GHLeafIssuesTest: XCTestCase {
     func testGH105() throws {
         do {
             let template = """
-            #if(1 + 1 == 2):hi#endif
+            #if((1 + 1) == 2):hi#endif
```
0xTim, line 53: This looks like we're changing the behaviour of the parser in a breaking way?

Tests/LeafKitTests/GHTests/VaporLeaf.swift
```diff
@@ -50,21 +50,14 @@ final class GHLeafIssuesTest: XCTestCase {
...
         do {
             let template = """
-            #if(2 == 1 + 1):hi#endif
-            """
-            let expected = "hi"
-            try XCTAssertEqual(render(template, ["a": "a"]), expected)
-        }
-        do {
-            let template = """
-            #if(1 == 1 + 1 || 1 == 2 - 1):hi#endif
```
0xTim, line 67: Why is this test removed?

0xTim (commented):
I've had a quick scan through and will do a more thorough review in the week - but to start it looks like a lot of the existing tests have been changed or commented out or just deleted - is this just while rewriting or are you planning on these changes being permanent?

Tests/LeafKitTests/GHTests/VaporLeaf.swift
pontaoski: yea, this and other minorly breaking changes are why i consider the PR leaf 5 material—some operators are deemed nonassociative and require you to add parentheses to avoid attempting to associate them.

tonyarnold (commented):
I'd love better error reporting in Leaf, so I'm really supportive of any efforts to achieve this @pontaoski :heart: Thanks for your explorations!

That being said, I've had a look through the proposed changes, and it's a large amount to change in one go. There could be an amazing set of improvements in here, but it's putting the onus on the reviewer to wade through thousands of lines of code and unstructured commits to work that out. 

Are you willing to consider breaking the PR into a series of smaller changes and improvements? Perhaps there is a logical sequence of individual improvements that could be separated out and proposed as individual PRs in a chain? 

I'm happy to help do this if you'd like.

Tests/LeafKitTests/GHTests/VaporLeaf.swift
tonyarnold: Why did you want to make operators non-associative, @pontaoski? Is there a technical limitation in the implementation, or was the change intentional?

pontaoski: intentional: from my experience, i've seen people struggle to comprehend operators like `==` and `!=` if used in an associative way, and there's usually no reason to chain multiple of them, so making the comparison operators nonassociative forces people to add parentheses for legibility in anything more complex than `a == b` (with an error making this explicit).

but changing this would be as simple as editing the table in the scanner where the comparison operators are defined as nonassociative.

tonyarnold: Thanks for answering my question 🎉 

Your explanation is reasonable, and I concur that the current syntax can be confusing.

This change seems unrelated to what you set out to achieve with the rest of this PR, though? It also seems pretty contained - this might be a good candidate for a separate PR that builds on the base improvements you've made.

Given that this change would break existing production code, it'd be worthwhile socialising the idea [via the Swift forums](https://forums.swift.org/c/related-projects/vapor/30) and giving users an opportunity to provide feedback. I'm happy to write this up as a forum thread if it's not something you're keen to do.

pontaoski: most of the PR is breaking changes; i honestly don't see why this one breaking change is being singled out.

tonyarnold: For me, it was because it was the easiest breaking change to see due to your new tests, and it was easy to see that it would affect users of Leaf (as opposed to developers working directly with LeafKit's public APIs). 

I would like to help get your changes cleaned up and merged, so if you're aware of other breaking changes that need to be handled in documentation and other consumers of this library, it would be helpful to know about them without having to go through the code line-by-line. Perhaps they could go in the PR description?

pontaoski:
in my experience regarding error reporting in parsers, you kinda need the capabilities to be worked in from the start of the design, otherwise redoing the whole thing from the ground up is much easier to understand, since you only need to validate that a brand new lexer+parser is doing the right thing is way easier than trying to validate that you've correctly hacked on better error reporting to an existing design.

additionally, unfortunately with the way leaf is designed, there was an extremely tight coupling between the data structures involved in the lexer+parser+renderer+evaluator, so any change of lexer+parser necessitates a lot of changes in the latter.

that being said, i've adjusted the file structure to make the git diff render all of the rewrites as new files instead of attempting to diff with the files they replace, which should help with legibility.

tonyarnold:
That makes sense. I also appreciate that you filed this PR so that these conversations could happen around how best to integrate your changes, so thank you for being open to that.  

To get this PR to a point where it can be properly reviewed and merged, I think there are a few things that need to happen: 

- Tidy up the PR description/title/etc - as you saw with #111, those get used in the release notes when PRs are merged. If you're confident in the changes you've made and their effect, I think you can describe them succinctly and clearly with that purpose in mind 👍🏻 
- Let @0xTim know via the inline review comment about the tests that have been modified, removed and commented out - I assume that's all just part of the churn of you exploring these changes.
- If you can, please separate out deliberate syntax changes and new features such as the requirement for parentheses and the nil boolean handling into separate PR(s). Correct me if I'm wrong, but your goal here was a rewrite of the parser/lexer/etc to provide a better design, with improved error reporting? If that's the case, let's focus this PR on just that, and move anything that doesn't work toward that goal into their own PRs. 

Sorry for the slow replies today, it's been a long work day.

Tests/LeafKitTests/GHTests/VaporLeaf.swift
pontaoski: moved it to `LeafParserTests.testGroupingTwo` to make it more clear that this is testing operator precedence

Tests/LeafKitTests/GHTests/VaporLeaf.swift
pontaoski: the comparison operators are associative now

pontaoski:
poke

PR #118, opened 2023-03-05
add isEmpty tag

fananek:
Adds a new `isEmpty` tag for checking empty Strings. Equivalent of Swift's `isEmpty`

Example:
```leaf
#if(isEmpty(emptyString)):
    This is an empty string.
#endif
```

### DIFF
--- a/Tests/LeafKitTests/TagTests.swift
+++ b/Tests/LeafKitTests/TagTests.swift
@@ -117,6 +117,21 @@ class TagTests: XCTestCase {
         """
         try XCTAssertEqual(render(template, ["core": ["Tanner", "Logan", "Gwynne", "Siemen", "Tim"]]), expected)
     }
+    
+    func testIsEmpty() throws {
+        let template = """
+        #if(isEmpty(emptyString)):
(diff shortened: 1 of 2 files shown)


Tests/LeafKitTests/TagTests.swift
```diff
@@ -117,6 +117,21 @@ class TagTests: XCTestCase {
         """
         try XCTAssertEqual(render(template, ["core": ["Tanner", "Logan", "Gwynne", "Siemen", "Tim"]]), expected)
     }
+    
+    func testIsEmpty() throws {
```
0xTim, line 121: Could you add a check in the test to ensure the inverse also works - so `isEmpty` not returning true? Otherwise good to go!

0xTim (changes requested):

Tests/LeafKitTests/TagTests.swift
fananek: Of course. See the latest commit.

0xTim (approved):
Thanks!

PR #119, opened 2023-03-06
Extend for loops with custom index support

vzsg:
Traditionally, the `for` loops in Leaf declare three variables in the loop's local context: `isFirst`, `isLast` and `index`. The first two aren't particularly interesting, but being able to access all indices in a *nested* for loop situation is sometimes useful when working with two dimensions at once. Not to mention it just popped up on Discord earlier today.

This PR extends Loop with the option of renaming `index`.

Example syntax:

```leaf
#for(i, array in arrays):
  #for(j, element in array):
    (#(i), #(j)): #(element)
  #endfor
#endfor
```

---

Should be semver-minor.


0xTim (approved):
Thank you for this!

Issue #122, opened 2023-04-10
Allow to add quotes into stringLiteral parameters


supakonoha:
**Is your feature request related to a problem? Please describe.**
I'm using a custom tag that has a string parameter. If this parameter includes a quote, this is interpreted as final quote for the parameter, although it's escaped. For example:

`#tag("value", "this is a \"custom\" parameter")`

for the second parameter, **LeafLexer** parses it as `this is a \` instead of `this is "custom" parameter` 

**Describe the solution you'd like**
I would like to allow quotes inside parameters using escape characters or other option.

roya1v:
As far as I see when we read a string we do not distinguish between quotes that were and were not escaped.
That's why I looked at escaping them using a different character and create a draft #124  of how it could work escaping it like `$\` but it should work with any other character too.
Let me know what you think!

supakonoha:
Looks nice!!!!! Thanks for this change!!!!

Just a suggestion, I don't know if we could define "$" as constant or add into enum with .quote?

PR #124, opened 2023-05-13
Allow escaped quotes in tag parameters

roya1v:
Allows escaped quotes in tag parameters so you can do something like `#myCustomTag("this \"thing\" is cool")`

Solves #122


roya1v:
@0xTim that was my initial idea. As far as I tried we don't distinguish between escaped and not-escaped quotes. I tried to dick deeper but couldn't find a place in leaf-kit to amend this behaviour. Correct me please if I'm wrong
Screenshot 2023-05-22 at 12 12 45
Screenshot 2023-05-22 at 12 13 02


Now that I think about it, two slashes might work, lmk if you would prefer that 
Screenshot 2023-05-22 at 19 34 22

roya1v:
Hey @0xTim sorry I don't know if you didn't see it or didn't have time to take a look. I've pushed the change to `\\`

PR #125, opened 2023-07-28
Full compatibility with `-strict-concurrency=complete`

MahdiBM:
These changes will make `leaf-kit` fully compatible with `-strict-concurrency=complete`.


gwynne (changes requested):
See comments, also I'm not really happy with the `SendableBox`, too many locks... but I don't know that I have any better suggestions. This package is easily the worst mess left in Vapor now that Fluent and some other things have been cleaned up some...

Sources/LeafKit/SendableBox.swift
```diff
@@ -0,0 +1,24 @@
+import NIOConcurrencyHelpers
+
+/// Uses a locking mechanism to ensure Sendability.
+internal final class SendableBox<Value>: @unchecked Sendable {
```
0xTim, line 4: Why are we not using the NIO helpers? This doesn't make un-Sendable objects Sendable unfortunately as I've learn't the hard way

Sources/LeafKit/LeafRenderer.swift
```diff
@@ -25,9 +25,12 @@ public final class LeafRenderer {
     public let sources: LeafSources
     /// The NIO `EventLoop` on which this instance of `LeafRenderer` will operate
     public let eventLoop: EventLoop
+    let _userInfo: SendableBox<[AnyHashable: Any]>
```
0xTim, line 28: This doesn't do what you think it does. The only way we can make `Any` `Sendable` is by tying this to an event loop

0xTim (changes requested):
We should add the flag as a step in CI too

Sources/LeafKit/SendableBox.swift
MahdiBM: What types are you thinking of? I don't think we can use types like NIOLoopBound since it requires an eventloop

MahdiBM: At least not everywhere. Like in global vars.

Sources/LeafKit/LeafCache/DefaultLeafCache.swift
```diff
@@ -1,11 +1,33 @@
...
+    /// Global setting for enabling or disabling the cache
+    public var isEnabled: Bool {
+        get {
+            self._isEnabled
+        }
+        set(newValue) {
+            self._isEnabled = newValue
+        }
+    }
     /// Current count of cached documents
```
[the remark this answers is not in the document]
MahdiBM: i assume compiler will take care of that anyway... two levels to not use `self.lock.withLock` 2 more times, which is also less code.

Sources/LeafKit/SendableBox.swift
MahdiBM: I forgot to add a `Value: Sendable` constraint. I assume that solves most of the problems?

0xTim: Use `NIOLockedValueBox` then rather than rolling our own. This will cause issues with `userInfo` as I'm sure you'll see

Sources/LeafKit/LeafRenderer.swift
MahdiBM: Makes sense ... even with a `Value: Sendable` constraint it is still showing warnings.
Not sure what exactly to do though.
We need to change `Any` -> `any Sendable` as well as somehow conforming `AnyHashable` to `Sendable`.

MahdiBM: Though not sure what an event loop does that the locking mechanism can't ... I'm tempted to ask in the OpenSource slack.

Sources/LeafKit/SendableBox.swift
MahdiBM: ah ... the usual problem of outdated comments.
yeah i didn't remember that type exists.

Sources/LeafKit/Exports.swift
```diff
@@ -2,8 +2,16 @@
...
+    private static let _tagIndicator = SendableBox(Character.octothorpe)
+    static var tagIndicator: Character {
+        get {
+            _tagIndicator.value
+        }
+        set(newValue) {
+            _tagIndicator.value = newValue
+        }
+    }
+
```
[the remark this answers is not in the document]
MahdiBM: ah didn't notice this one considering i was rushing everything.
Though still we need the mutability.

MahdiBM:
@gwynne should the reusable CI has an option for strict concurrency checking? (automatically set to `complete`, or with the option for users to set the flag to values like `targeted`/`complete`)

Sources/LeafKit/LeafRenderer.swift
MahdiBM: The documentation of `NIOLoopBound` does mention why...

Sources/LeafKit/LeafRenderer.swift
```diff
@@ -25,9 +26,12 @@ public final class LeafRenderer {
     public let sources: LeafSources
     /// The NIO `EventLoop` on which this instance of `LeafRenderer` will operate
     public let eventLoop: EventLoop
+    let _userInfo: NIOLoopBound<[AnyHashable: Any]>
     /// Any custom instance data to use (eg, in Vapor, the `Application` and/or `Request` data)
-    public let userInfo: [AnyHashable: Any]
-    
+    public var userInfo: [AnyHashable: Any] {
+        _userInfo.value
```
MahdiBM, line 32: Will this crash on access by users on another `eventLoop`?

MahdiBM:
`NIOLoopBound` crashing in CI .... not sure what to do with the dictionary.

MahdiBM:
It appears for a dictionary like that, Foundation itself just uses `@unchecked Sendable` while using a lock and also the `_modify` accessor (not sure if `_modify` help for Sendability at all):
https://github.com/apple/swift-foundation/blob/90d50e8cbfa9ba002fde3df3839a5bbd8b33b402/Sources/FoundationEssentials/JSON/JSONEncoder.swift#L303
https://github.com/apple/swift-foundation/blob/90d50e8cbfa9ba002fde3df3839a5bbd8b33b402/Sources/FoundationEssentials/JSON/JSONEncoder.swift#L1379

0xTim:
Err @fabianfett that's wrong right?

@MahdiBM the issue is if you store something that's not `Sendable` in the dictionary and store a reference to it then something outside of the context can mutate it at the same time and so 💥

MahdiBM:
Yeah it's suboptimal, but it's something 😅

0xTim:
Nah it's really bad as Fabian explained to me 😅 essentially you go from the compiler giving you a warning saying this might be a problem to the user to the compiler completely removing that warning so the end user never knows. I'd rather have a dictionary that takes a Sendable type and pass the warning on to the user

Issue #126, opened 2023-12-07
`#extend` with two parameters doesn’t pass second parameter as context when nested


benblakely:
### Description

When nesting an `#extend` with two parameters, the second parameter seems to be ignored instead of being provided as context to the template. 

### To Reproduce

Here’s a test (based on the existing `testExtendWithSugar`) that shows the issue and fails:

```
func testNestedExtendWithSugar() throws {
    let layout = """
    <body>#import("main")</body>
    """

    let header = """
    <h1>#(child)</h1>
    """

    let base = """
    #extend("layout"):#export("main"):#extend("header", parent)#endexport#endextend
    """

    let expected = """
    <body><h1>Elizabeth</h1></body>
    """

    let layoutAST = try LeafAST(name: "layout", ast: parse(layout))
    let headerAST = try LeafAST(name: "header", ast: parse(header))
    let baseAST = try LeafAST(name: "base", ast: parse(base))

    let baseResolved = LeafAST(from: baseAST, referencing: ["layout": layoutAST, "header": headerAST])

    var serializer = LeafSerializer(
        ast: baseResolved.ast,
        ignoreUnfoundImports: false
    )
    let view = try serializer.serialize(context: ["parent": ["child": "Elizabeth"]])
    let str = view.getString(at: view.readerIndex, length: view.readableBytes) ?? ""

    XCTAssertEqual(str, expected)
}
```

The test fails with: 

```
XCTAssertEqual failed: ("<body><h1></h1></body>") is not equal to ("<body><h1>Elizabeth</h1></body>")
```

### Expected behavior

My understanding from #111 is that when `#extend` is given two parameters, the second parameter should be passed as context to the template (which sounds like a very useful feature). That behavior seems to happen correctly when the `#extend` isn’t nested, but doesn’t work when nested.

My hope is to use this feature to pass data around like the example below:

items/list.leaf:
```
#extend("layouts/main"):
  #export("content"):
    <h1>Items</h1>
    #for(item in items):
      #extend("items/item", item)
    #endfor
  #endexport
#endextend
```

items/item.leaf:
```
<div>
  <h2>#(name)</h2>
  …
</div>
```

### Environment

* Vapor Framework version: 4.85.1
* Vapor Toolbox version: 18.7.4
* OS version: macOS 14.1.1 (23B81)

### Additional context

I’ve tried to figure out how to fix the issue but haven’t been able to yet.

[...]

PR #127, opened 2023-12-08
Fix nested extend with sugar

benblakely:
When nesting an `#extend` with two parameters, the second parameter was ignored instead of being used as context.

Fixes vapor/leaf-kit#126


gwynne (approved):

gwynne:
Looks good, thanks!!

Issue #128, opened 2023-12-08
`#extend` not rendered when wrapped by `#with`


benblakely:
### Describe the bug

When a `#with` wraps an `#extend`, the `#extend` doesn’t seem to render. (I mentioned this issue when I reported #126 but they appear to be separate issues.)

### To Reproduce

Here’s a test that shows the issue and currently fails:

```
func testWithWrappingExtend() throws {
    let header = """
    <h1>#(child)</h1>
    """

    let base = """
    <body>#with(parent):<main>#extend("header")</main>#endwith</body>
    """

    let expected = """
    <body><main><h1>Elizabeth</h1></main></body>
    """

    let headerAST = try LeafAST(name: "header", ast: parse(header))
    let baseAST = try LeafAST(name: "base", ast: parse(base))

    let baseResolved = LeafAST(from: baseAST, referencing: ["header": headerAST])

    var serializer = LeafSerializer(
        ast: baseResolved.ast,
        ignoreUnfoundImports: false
    )
    let view = try serializer.serialize(context: ["parent": ["child": "Elizabeth"]])
    let str = view.getString(at: view.readerIndex, length: view.readableBytes) ?? ""

    XCTAssertEqual(str, expected)
}
```

This test fails with:

```
XCTAssertEqual failed: ("<body><main></body>") is not equal to ("<body><main><h1>Elizabeth</h1></main></body>")
```

### Expected behavior

The `#with` call should provide its parameter as context to the `#extend`.

### Environment

* Vapor Framework version: 4.85.1
* Vapor Toolbox version: 18.7.4
* OS version: macOS 14.1.1 (23B81)

### Additional context

I think I see what the problem is. I’ll submit a PR shortly.

PR #129, opened 2023-12-08
Fix extend not rendered when wrapped by with

benblakely:
When an `#extend` was wrapped by a `#with`, the `#extend` wasn't rendered.

Fixes vapor/leaf-kit#128


gwynne:
This looks fine at a glance, but I'd feel better if you could add another test or two that shows (or at least tries to show, since you can't necessarily prove a negative) that this change doesn't accidentally allow otherwise invalid syntax. (I can't claim to have dealt with the LeafKit code anywhere near recently enough to be able to suggest anything more specific than that, unfortunately 😕. @0xTim might, though.)

0xTim (approved):
Thanks for this, I think we're good here as it was mainly an oversight when the `with` tag was added

PR #130, opened 2024-03-04
Fix crash with missing parameter

b-nassler:
This line https://github.com/vapor/leaf-kit/blob/main/Sources/LeafKit/LeafParser/LeafParser.swift#L203
would cause a crash when no parameters are found

Fixed issue that lead to a crash when parameters would be empty in line https://github.com/vapor/leaf-kit/blob/main/Sources/LeafKit/LeafParser/LeafParser.swift#L203

### DIFF
--- a/Sources/LeafKit/LeafError.swift
+++ b/Sources/LeafKit/LeafError.swift
@@ -32,7 +32,7 @@ public struct LeafError: Error {
         /// - Provide template name & ordered array of template names that causes the cycle path
         case cyclicalReference(String, [String])
         /// Parameter missing
-        case missingParameter(String)
+        case missingParameter
 
         // MARK: Wrapped Errors related to Lexing or Parsing
         /// Errors due to malformed template syntax or grammar
@@ -86,8 +86,8 @@ public struct LeafError: Error {
                 return "\(src) - \(key) cyclically referenced in [\(chain.joined(separator: " -> "))]"
             case .lexerError(let e):
                 return "Lexing error - \(e.localizedDescription)"
-            case .missingParameter(let message):
-                return message
+            case .missingParameter:
+                return "Expected a parameter but found none"
         }
     }
     
--- a/Sources/LeafKit/LeafError.swift
+++ b/Sources/LeafKit/LeafError.swift
@@ -31,6 +31,8 @@ public struct LeafError: Error {
         /// Attempt to render an AST with cyclical external references
         /// - Provide template name & ordered array of template names that causes the cycle path
         case cyclicalReference(String, [String])
+        /// Parameter missing
+        case missingParameter(String)
 
         // MARK: Wrapped Errors related to Lexing or Parsing
         /// Errors due to malformed template syntax or grammar
@@ -84,6 +86,8 @@ public struct LeafError: Error {
                 return "\(src) - \(key) cyclically referenced in [\(chain.joined(separator: " -> "))]"
             case .lexerError(let e):
                 return "Lexing error - \(e.localizedDescription)"
+            case .missingParameter(let message):
+                return message
         }
     }
     
--- a/Sources/LeafKit/LeafError.swift
+++ b/Sources/LeafKit/LeafError.swift
@@ -31,8 +31,6 @@ public struct LeafError: Error {
         /// Attempt to render an AST with cyclical external references
         /// - Provide template name & ordered array of template names that causes the cycle path
         case cyclicalReference(String, [String])
-        /// Parameter missing
-        case missingParameter
 
         // MARK: Wrapped Errors related to Lexing or Parsing
         /// Errors due to malformed template syntax or grammar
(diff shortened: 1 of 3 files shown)


Sources/LeafKit/LeafParser/LeafParser.swift
```diff
@@ -200,7 +200,13 @@ internal struct LeafParser {
                     let params = try readParameters()
                     // parameter tags not permitted to have bodies
                     if params.count > 1  { group.append(.expression(params)) }
-                    else { group.append(params.first!) }
+                    ///// else { group.append(params.first!) }
```
gwynne, line 203: Can we just delete this line entirely?
```suggestion
```

Sources/LeafKit/LeafParser/LeafParser.swift
```diff
@@ -200,7 +200,13 @@ internal struct LeafParser {
...
+                    else {
+                      guard let firstParam = params.first else {
+                        throw LeafError(.missingParameter("Expected a parameter but found none"))
```
gwynne, line 206: I would prefer that this message be returned from `LeafError.localizedDescription`, and that the new error case not have the associated string at all.
```suggestion
                        throw LeafError(.missingParameter)
```

Sources/LeafKit/LeafError.swift
```diff
@@ -84,6 +86,8 @@ public struct LeafError: Error {
                 return "\(src) - \(key) cyclically referenced in [\(chain.joined(separator: " -> "))]"
             case .lexerError(let e):
                 return "Lexing error - \(e.localizedDescription)"
+            case .missingParameter(let message):
+                return message
```
gwynne, line 90:
```suggestion
                return "Expected a parameter but found none"
```
As per my comment in `LeafParser.swift`.

Sources/LeafKit/LeafParser/LeafParser.swift
```diff
@@ -200,7 +200,13 @@ internal struct LeafParser {
                     let params = try readParameters()
                     // parameter tags not permitted to have bodies
                     if params.count > 1  { group.append(.expression(params)) }
-                    else { group.append(params.first!) }
+                    ///// else { group.append(params.first!) }
+                    else {
+                      guard let firstParam = params.first else {
```
gwynne, line 205: Can you fix this in the `dump()` function on line 187 as well? It's fine to make it throwing in the process.

gwynne (changes requested):
Nice catch and great fix! Just a couple of nitpicks, see the inline comments 👍

gwynne:
Oh, I forgot to also add - can you add a test for this as well?

Sources/LeafKit/LeafParser/LeafParser.swift
b-nassler: Makes sense!

Sources/LeafKit/LeafParser/LeafParser.swift
b-nassler: done

b-nassler:
@gwynne Updated, please check 🙂

b-nassler:
@gwynne @0xTim Everything should be ok I think. Let me know if I should do anything else. Thanks!

Tests/LeafKitTests/LeafErrorTests.swift
```diff
@@ -45,4 +45,26 @@ final class LeafErrorTests: XCTestCase {
...
+        switch error.reason {
+        case .missingParameter:
+          return
+        default:
+          XCTFail("Wrong error: \(error.localizedDescription)")
+        }
+      } catch {
+        XCTFail("Wrong error: \(error.localizedDescription)")
+      }
+    }
```
gwynne, line 69: This can be replaced by:
```suggestion
        XCTAssertThrowsError(try TestRenderer(sources: .singleSource(test))
            .render(path: "missingParam", context: [:])
            .wait()
        ) {
            guard case .missingParameter = ($0 as? LeafError)?.reason else {
                return XCTFail("Expected LeafError.missingParameter, got \(String(reflecting: $0))")
            }
        }
```

gwynne (changes requested):
Great! One last nit and all should be good to go 🙂

b-nassler:
@gwynne Good one! I committed your suggestion

Sources/LeafKit/LeafError.swift
```diff
@@ -31,6 +31,8 @@ public struct LeafError: Error {
         /// Attempt to render an AST with cyclical external references
         /// - Provide template name & ordered array of template names that causes the cycle path
         case cyclicalReference(String, [String])
+        /// Parameter missing
+        case missingParameter
```
0xTim, line 35: Ok I'm going to come in and ruin the party, but only a little - we can't add a new enum case without it being a breaking change unfortunately. And the LeafError is not really one of the enums that we consider 'private'. The quick fix to this will be to throw one of the existing errors, probably `unknownError`. That will allow us to release it as a patch version

0xTim (changes requested):

Sources/LeafKit/LeafError.swift
b-nassler: And we had so much fun here. Oh well have it your way then ;-)
Updated as per request.

0xTim (approved):
Thank you!

b-nassler:
@0xTim You probably saw this but something ran out of memory so it didn't succeed merging

Issue #123, opened 2023-04-20
`requireBody()` and `requireNoBody()` not working as expected 


ptoffy:
### Describe the bug

Whilst creating a custom `UnsafeUnescapedLeafTag`, `context.requireNoBody()` always throws even if no body is provided, while `context.requireBody()` never throws even when no body is provided, rather returning an empty `[Syntax]`. 

While looking through the tests I noticed these methods are not tested anywhere and I think simply checking for the array being empty in addition to checking for it being `!= nil` could solve the issue. The other option would be searching where the `body` of the tag is passed in and setting it to `nil` if it's an empty array

### To Reproduce

1. Create a simple project just using Leaf 4
2. Create 
```swift
struct BodyRequiringTag: UnsafeUnescapedLeafTag {
    func render(_ ctx: LeafContext) throws -> LeafData {
        _ = try ctx.requireBody()
        
        return .string("Hello there")
    }
}
```
and 
```swift
struct NoBodyRequiringTag: UnsafeUnescapedLeafTag {
    func render(_ ctx: LeafContext) throws -> LeafData {
        try ctx.requireNoBody()
        
        return .string("General Kenobi")
    }
}
```
3. Add the tags to the test project
```swift
app.leaf.tags["bodytag"] = BodyRequiringTag()
app.leaf.tags["nobodytag"] = NoBodyRequiringTag()
```

4. Adding the first tag to the project
```
#bodytag:#endbodytag
```
or 
```
#bodytag()
```
This returns `Hello there`, which is just the return type of the tag, while it should be returning `{error: true, reason: "Missing body"`

5. Adding the other tag
```
#nobodytag
```
This returns `{error: true, reason: "Extraneous body"}`
The only way to make this error go away is to set the template as
```
#nobodytag:#endnobodytag
```
Which is, I think, not that intuitive

### Environment

* Vapor Framework version: `4.76.0`
* Vapor Toolbox version: `18.6.0`
* OS version: `MacOS Ventura 13.0`
* Xcode version: `14.1 (14B47b)`

ptoffy:
Closed by #133

PR #133, opened 2024-08-31
Fix and add tests for `requireBody()` and `requireNoBody()`

fpseverino:
### Issue #123 

- Fix and add tests for `requireBody()` and `requireNoBody()`


0xTim (approved):
Thanks!

PR #134, opened 2025-01-18
Update to Swift 5.9

fpseverino:
- Update `swift-tools-version` to 5.9
- Add common `swiftSettings` to targets
- Fix unhandled files warnings
    - `Templates` folder in test target; added to target's resources
    - `Docs.docc` folder in main target; updating `swift-tools-version` fixes it
- `Lock` was deprecated in favour of `NIOLock`
- Add MUSL and iOS in CI

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.8
 import PackageDescription
 
 let package = Package(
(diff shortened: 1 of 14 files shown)


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.8
```
fpseverino, line 1: The `Docs.docc` problem is solved with any version higher than 5.4, I put 5.8 only because `vapor/leaf` is on 5.8

gwynne: The correct minimum to use at this point in time is 5.9.

gwynne: Ideally you would also enable various Swift settings (most especially strict concurrency), but that would likely entail a good bit of work to clear up `Sendable` issues.

fpseverino: Ok, I have added the common 5.9 Swift settings, the only one that required immediate changes is "ExistentialAny", strict concurrency just gives a bunch of warnings.
Should we keep the setting and warnings or should I disable it for now?

fpseverino: [This PR](https://github.com/vapor/leaf-kit/pull/125) could be revisited

0xTim (approved):

Package.swift
0xTim: Keep the warnings for now (they shouldn't appear in adopters packages I don't _think_) and then we can follow up with #125

PR #135, opened 2025-04-01
Sendable fixes and other cleanup

gwynne:
**These changes are now available in [1.13.0](https://github.com/vapor/leaf-kit/releases/tag/1.13.0)**

The big changes in this PR are making the code `Sendable`-clean and minimally strict Concurrency-correct (emphasis on "minimally", there's still a bunch of stuff with futures going on). Accordingly with the release of Swift 6.1, the package also now requires a minimum version of Swift 5.10.

Additional changes:
- README and docs appearance updated.
- Enabled Android CI.
- Replaced the convoluted "performant" HTML escaping logic with straightforward string replacement calls which turn out to actually be more performant than the circa Swift 4.2 unsafe pointer stuff.
- Fixed the tests so they no longer fail if run on a machine set to a timezone where DST is currently in effect. Tests also no longer use `EventLoopFuture.wait()`.
- A pass across the board for basic style, formatting, and readability. There's a bit less force-unwrapping going around for good measure.


gwynne:
The failure of the integration tests is spurious; the `SWIFT_DETERMINISTIC_HASHING` requirement is removed by vapor/leaf#237.

Sources/LeafKit/String+HTMLEscape.swift
```diff
@@ -1,103 +1,25 @@
+#if swift(<6.0)
 import Foundation
```
ptoffy, line 2: Maybe dance the `FoundationEssentials` dance wherever it's used?

Tests/LeafKitTests/TagTests.swift
```diff
@@ -175,7 +175,11 @@ class TagTests: XCTestCase {
         The date is 2020-11-09T14:30:00
         """
 
-        let now = 1604932200 - Calendar.current.timeZone.secondsFromGMT()
+        let formatter = DateFormatter()
```
ptoffy, line 178: Not really necessary but I'd cache the formatter

ptoffy (approved):
Overall this LGTM, I'd add Vapor's `.swift-format` file and formatting CI given all the style changes

Sources/LeafKit/String+HTMLEscape.swift
gwynne: Then I'd start having to go changing all the `import NIO`s to `import NIOCore` and such too, that takes effort 😆

Tests/LeafKitTests/TagTests.swift
gwynne: For the purposes of a single use by each of four tests? Hardly worth it.

Sources/LeafKit/String+HTMLEscape.swift
gwynne: There, I did both. Happy? 😁

0xTim (approved):
Couple of minor nits, nothing major

